PR #26, opened 2024-02-23
Project01 - String Library Implementation

emuwang:
Implemented the following functions:
- char *strncpy(char *dst, const char *src, uint n);
- int strncmp(const char *s1, const char *s2, uint n);
- char *strncat(char *restrict s1, const char *restrict s2, uint n);
- char *strstr(const char *haystack, const char *needle);
- uint strspn(const char *s, const char *accept);
- uint strcspn(const char *s, const char *reject);
- char *strtok(char *restrict str, const char *restrict sep);
- char *strtok_r(char *restrict str, const char *restrict sep, char **restrict last);
- char *strsep(char **stringp, const char *delim);
- int scanf(const char *restrict format, ...);


sghahghahi:
I will code review this!

atanaspatterson:
I'll also code review it!!

atanaspatterson:
Hi all, here's my suggested changes/improvements:
- in scanf, I tested passing strings instead of integers, and they still got converted into integers leading to confusing output. For this, maybe comparing the input strings with the result given by atoi can help. Maybe even simpler logic could work, since I see that you pretty much show length 0, so  you can maybe print something else if your length == 0. In addition, the second part did not seem to work well in most cases:
`scanf:`
`Input two integers(space separated, can be negative): a b`
`0 0 0`
`Input three strings(space separated): hi hello hi`
`1 hi (null)`
- Passing NULL to the string functions will break them. I tested passing it in strspn and strcat, both of which broke the code (used my own test cases file). Probably just handling NULL cases at the start of each function would solve this (like 
`if str1 == NULL || str2 == NULL) { handle } `

- `-Testing strncat with null:`
`Concatenated string: Hello, i�"�& (weird output)` 
`Testing strcspn with null:`
`Span of characters: 10 (passed null - should not be 10`
- For one more tiny thing, my compiler (for some reason) did not like `strtok_r` being called inside `strtok` before it was defined, so simply swapping the function definitions would do the trick(strtok_r before strtok).

The string functions look great - took me some time to find anything that was wrong. Also, the man pages and the provided test file were perfect, you did great with this project version!

Jsk07211:
> Hi all, here's my suggested changes/improvements:
> 
>     * in scanf, I tested passing strings instead of integers, and they still got converted into integers leading to confusing output. For this, maybe comparing the input strings with the result given by atoi can help. Maybe even simpler logic could work, since I see that you pretty much show length 0, so  you can maybe print something else if your length == 0. In addition, the second part did not seem to work well in most cases:
>       `scanf:`
>       `Input two integers(space separated, can be negative): a b`
>       `0 0 0`
>       `Input three strings(space separated): hi hello hi`
>       `1 hi (null)`
> 
>     * Passing NULL to the string functions will break them. I tested passing it in strspn and strcat, both of which broke the code (used my own test cases file). Probably just handling NULL cases at the start of each function would solve this (like
>       `if str1 == NULL || str2 == NULL) { handle } `
> 
>     * `-Testing strncat with null:`
>       `Concatenated string: Hello, i�"�& (weird output)`
>       `Testing strcspn with null:`
>       `Span of characters: 10 (passed null - should not be 10`
> 
>     * For one more tiny thing, my compiler (for some reason) did not like `strtok_r` being called inside `strtok` before it was defined, so simply swapping the function definitions would do the trick(strtok_r before strtok).
> 
> 
> The string functions look great - took me some time to find anything that was wrong. Also, the man pages and the provided test file were perfect, you did great with this project version!

Thanks for the code review! With regards to scanf and the integers, it actually returns zero because we used strcspn, so if it encounters any non-digit character (except if it starts with '-' for negative numbers), in this case the length will be a 0, we atoi 0, hence the result :) The second test case for scanf string uses: initialisedbuf0, uninitialisedbuf, initialisedbuf1. Since the first buffer is initialised, it will save the value, while the uninitialised buffer returns null and stops the code, so the third string doesn't actually get read!

With regards to strspn and strcat and null strings (if I understand your test correctly), they simply seg fault in the standard C String library. For strncat, this could be because it relies on the programmer to correctly input the size of the destination buffer, but thank you for the wonderful suggestion! We will definitely add it to our implementation just in case :) 

Will take note of the strtok_r and strtok issue as well! Thank you for the comments!!

Jsk07211:
> > Hi all, here's my suggested changes/improvements:
> > ```
> > * in scanf, I tested passing strings instead of integers, and they still got converted into integers leading to confusing output. For this, maybe comparing the input strings with the result given by atoi can help. Maybe even simpler logic could work, since I see that you pretty much show length 0, so  you can maybe print something else if your length == 0. In addition, the second part did not seem to work well in most cases:
> >   `scanf:`
> >   `Input two integers(space separated, can be negative): a b`
> >   `0 0 0`
> >   `Input three strings(space separated): hi hello hi`
> >   `1 hi (null)`
> > 
> > * Passing NULL to the string functions will break them. I tested passing it in strspn and strcat, both of which broke the code (used my own test cases file). Probably just handling NULL cases at the start of each function would solve this (like
> >   `if str1 == NULL || str2 == NULL) { handle } `
> > 
> > * `-Testing strncat with null:`
> >   `Concatenated string: Hello, i�"�& (weird output)`
> >   `Testing strcspn with null:`
> >   `Span of characters: 10 (passed null - should not be 10`
> > 
> > * For one more tiny thing, my compiler (for some reason) did not like `strtok_r` being called inside `strtok` before it was defined, so simply swapping the function definitions would do the trick(strtok_r before strtok).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The string functions look great - took me some time to find anything that was wrong. Also, the man pages and the provided test file were perfect, you did great with this project version!
> 
> Thanks for the code review! With regards to scanf and the integers, it actually returns zero because we used strcspn, so if it encounters any non-digit value (excluding if it starts with a '-' for negative numbers), in this case the length will be a 0, we atoi 0, hence the result :) The second test case for scanf string uses: initialisedbuf0, uninitialisedbuf, initialisedbuf1. Since the first buffer is initialised, it will save the value, while the uninitialised buffer returns null and stops the code, so the third string doesn't actually get read!
> 
> With regards to strspn and strcat and null strings (if I understand your test correctly), they simply seg fault in the standard C String library. For strncat, this could be because it relies on the programmer to correctly input the size of the destination buffer, but thank you for the wonderful suggestion! We will definitely add it to our implementation just in case :)
> 
> Will take note of the strtok_r and strtok issue as well! Thank you for the comments!!

sghahghahi:
You both did a very nice job incorporating useful string functions into xv6. @atanaspatterson made some great observations that I'd like to add to. Here's one vulnerability I found in `scanf()`:

`scanf()` is easy to break with a buffer overflow. We get a `usertrap()` if the user input is significantly larger than the buffer size
- A useful addition to `scanf()` might be a check for a field width specifier like this:
- ```c
   char buf[100];
   scanf("%99s", buf); // Only read 99 characters
   ```

Again, well done with this project and thank you for adding some much-needed (and well-documented) string functions to xv6!

user/str.c
```diff
@@ -0,0 +1,129 @@
...
+
+	// strtok
+	printf("\nstrtok:\n");
+    char strtokstr[] = "first second third";
+    char *strtoktoken = strtok(strtokstr, " ");
+    printf("%s\n", strtoktoken);
+
+	// strtok_r
+	printf("\nstrtok_r:\n");
+	char str[] = "Hello, World! Geeks for Geeks.";
```
malensek: Seems like this was lifted from https://www.geeksforgeeks.org/strtok-strtok_r-functions-c-examples/ ?

Can you check what the license is on that code sample? If there's no license this test should be removed and replaced with something you've implemented or an open source test.

user/str.c
```diff
@@ -0,0 +1,129 @@
...
+	len = strcspn(buf, "Hel");
+	printf("len: %d, should be 0\n", len);
+
+	len = strcspn(buf, "hel");
+	printf("len: %d, should be 1\n", len);
+
+	len = strcspn(buf, " ");
+	printf("len: %d, should be 5\n", len);
+
+	// strtok
```
malensek, line 66: Doesn't look like this tests more than a single call to strtok. Based on the implementation, I think it may only work once? There should be a global / static variable holding the state of strtok.

user/string.c
```diff
@@ -0,0 +1,158 @@
...
+	} else { // terminate the token with null terminator point last to 1 space beyond it
+		*end = '\0';
+		*last = end + 1;
+	}
+	return str;
+}
+
+char
+*strtok(char *restrict str, const char *restrict sep) {
+	char *last;
```
malensek: Perhaps making this static will work? It seems that currently this function will only work once unless I'm missing something.

user/scanf.c
```diff
@@ -0,0 +1,98 @@
...
+					buf += offset;
+					success++;	
+				}
+			} 
+			else if (c == 's') {
+				s = va_arg(ap, char *);
+
+				// No buffer space allocated, not successfully copied
+				if (s == 0) {
+					s = malloc(8);
```
malensek: Will this leak?

malensek (commented):
Documentation looks fantastic. Tests are very minimal -- I'd like to see more here but it's not something you need to do to finish P1. I made a few comments for you to check out and address; once those have been addressed you're all done!

user/scanf.c
Jsk07211: Hi Professor,
I removed those three lines and tested it, so now the code shouldn't lead to a memory leak. It was safe to remove since printf handles for null buffers, and allocating the buffer should be done by the user. Apologies.

user/str.c
emuwang: Will fix this, thank you for pointing it out!

user/str.c
emuwang: Sorry for not testing this! Changing last in strtok to be static fixed this. Thank you!

user/string.c
emuwang: Changed last to static as suggested. It should be working for multiple calls now. Thank you for the suggestion!

user/str.c
emuwang: Doesn't seem like there was a license for it. Changed it to an original test case to avoid a lawsuit.

allureking:
### **Memory Leak Analysis from Honghuai Ke for this String Library Implementation Project**
**Leak Check Report**
```
-- Leak Check --
[BLOCK 0x0000000000006140] 160  getline
-- Summary --
160 blocks lost (4096 bytes)
```
I conducted a thorough memory leak check on the string library implementation, focusing on the `getline` function, which was identified as a potential source of leaks. Here are the results from my own `malloc_leaks()` utility:
**Leak Detected**: 160 blocks totaling 4096 bytes
**Location**: Allocations made within the `getline` function


**Findings**
The leak check revealed that the `getline` function does not properly free memory that it allocates during execution. This has resulted in a significant memory leak, as evidenced by the memory not being released back to the system.

**Suggested Solutions**

**1. Review and Refactor getline:**
- The function should be carefully reviewed to ensure that all allocated memory is freed correctly before the function exits. This may involve adding specific `free()` calls at points where memory is no longer needed.
- Consider implementing a cleanup section within the function to handle all error paths and ensure memory is always freed.

**2. Implement Comprehensive Unit Tests:**
- Develop unit tests that specifically check for memory allocation and deallocation within `getline`.
- Use the `malloc_leaks()` utility as part of regular testing to catch any future regressions.

Jsk07211:
One of our test cases for scanf tested for a null buffer, which meant free(copy) didn't run. It has since been fixed :D
Screenshot 2024-05-13 at 1 24 50 PM

allureking:
> One of our test cases for scanf tested for a null buffer, which meant free(copy) didn't run. It has since been fixed :D Screenshot 2024-05-13 at 1 24 50 PM

I was just about to add to my previous commment that after further investigating, the memory allocated by getline is actually not being properly freed by scanf, which should be responsible for the lifecycle of this memory. Below are the two places of where free should be added:

At line 18: Add free(copy); after the completion of its usage.
At line 77: Ensure free(copy); is added to handle all exit points where getline memory might not be freed.

I saw that you've already fixed it, great job!

PR #27, opened 2024-02-23
Add uniq command

katherineanthony:
Vinnie and I completed the implementation for uniq as follows:
- flag of -c prints the count of the unique lines
- flag of -w prints the unique words
- flag of -wc prints the unique words and their counts


aajoseph2:
HI, I can code review this one!

aajoseph2:
Proposed Enhancements

    Case Sensitivity:
        Implement functionality to treat words with different cases as the same (e.g., "Blah" and "blah" should be considered identical, with only "blah" being outputted).

    Ignoring Trailing and Leading Spaces:
        Add functionality to ignore leading and trailing spaces (e.g., " blah" and "blah" should be treated as the same, with only "blah" being outputted).

Test Input and Observations

    Using the input below with the -w flag, a user trap error is encountered, suggesting the program does not correctly parse blank spaces:
        blah blah blah
        test
        test
        test
        test
        Blah
               blah

Inefficient and Potentially Problematic Sorting Algorithm

    The use of bubble sort may be inefficient for large datasets, and its implementation does not account for potential issues with duplicate or very similar strings, which could lead to an incorrect sort order in edge cases.

Identified Issues

    Memory Management:
        In the lines() and words() functions, memory is allocated but not freed, potentially leading to memory leaks.

    Error Handling in File Operations:
        There is no explicit error handling after file operations like open or close. Failure to open files or read/write errors could lead to undefined behavior or crashes.

    Logic in print_uniq():
        The comparison logic assumes the existence of lines[curr + 1] without verifying that curr + 1 < count. This could lead to accessing out-of-bounds memory, resulting in undefined behavior or crashes.

    Increased Size of File Input Text:
        "I cannot tell if it is just my machine, but when I double the lines of README.md, trying to read more lines or words, the program seems to enter an infinite loop and never finds the EOF. Also, when adding spaces to certain lines and words, execution tends to lead to undefined behavior."

Example Test Input That

    test file
    test
    testing this file
    test file
    test file
    blah blah blah
    blah
    hello
    joe
    bat
    okay
    what
    are
    we
    doing
    here
    again

Overall

"I had a lot of fun playing with the code and program. Thanks for allowing me to peer-review the code. Also, please let me know if I misinterpreted or misjudged something from the program's implementation."

AndrewLiu666:
Hi, Hana and I ( Andrew ) will code review this project!

hanag1234:
Hi, great job on your project! 
A few suggestions we (Andrew and Hana) would suggest are: 

1. Memory Management:  In the `lines` function, memory allocated with `malloc` for `line` is immediately freed after being added to the `lines` array. Instead, I suggest you free the memory only after you're done using the allocated strings, which should be after `print_uniq` has processed them.

2. Efficiency Concerns: The use of bubble sort for sorting lines or words is not the most efficient choice, especially for larger datasets. Consider implementing or using a more efficient sorting algorithm, like quicksort or mergesort, to improve performance.

3. Logic Error: The loop in `print_uniq` does not correctly handle the last unique word or line since it relies on comparing the current item with the next one, which doesn't work for the last element.

4. Incorrect Continuation in Loop in `print_uniq`: The `else if` condition checking for `"\0"` and `"\n"` uses `continue` without incrementing `curr`, leading to an infinite loop.

Overall, I enjoyed using the command and thought you did a great job.

user/ulib.c
```diff
@@ -145,3 +201,216 @@ memcpy(void *dst, const void *src, uint n)
...
+    memcpy(temp, line, sz);
+    free(line); // free old memory
+   	lines[count] = temp;
+   	line = temp;
+   	count++;
+  }	
+	
+	bubble_sort(lines, count); // sort all the lines
+  print_uniq(count, lines, cflag); // print uniq instances
+	free(lines);
```
malensek: lines is allocated on the stack, shouldn't free it here

user/ulib.c
```diff
@@ -145,3 +201,216 @@ memcpy(void *dst, const void *src, uint n)
...
+      break;
+    }
+    words[wordcount] = word;
+    wordcount++;
+  }
+
+  close(read);
+	bubble_sort(words, wordcount); // sort all the words
+  print_uniq(wordcount, words, cflag); // print unique words
+	free(words);
```
malensek: words is allocated on the stack, should not free it

user/ulib.c
```diff
@@ -145,3 +201,216 @@ memcpy(void *dst, const void *src, uint n)
...
+* @param fd file descriptor of file to read
+* @param cflag boolean for the -c flag
+*/
+void lines(int fd, bool cflag) {
+	int sz = 20; // getline will resize if necessary
+	int count = 0; // num of lines
+  char *lines[1024];
+
+	while (true) {
+		char *line = malloc(sz);
```
malensek: getline should automatically allocate this if sz = 0 and line = NULL

user/ulib.c
malensek: hmm.... but each element inside lines IS on the heap... so they all need to be freed to prevent memory leaks. (same goes for the other instance of this, below)

malensek (commented):
I made some notes on some memory issues here. There is one larger refactor I want you to do: move the code for this utility to uniq.c. You generally wouldn't see functions like this in the libc; they should live in the actual command line program's implementation instead. To finish P1:
* Move implementation out of ulibc
* Fix memory errors
* There is no documentation! (/docs)

FUTURE WORK (not required for P1): 
* we have a hard-coded limit of 1024 lines, that's something that should certainly be addressed. 
* use of a temp file is fine, but this could actually be achieved with pipes instead, so we never have to touch the FS.

katherineanthony:
@malensek Vincent and I have completed these code corrections. Thank you!

PR #29, opened 2024-02-26
Added Ls in FogOS!

ajpolintan:
Added colors to ls function for FogOs with the following flags:
    -t : list the type of files in a directory
    -l : list a long listing of files in a directory
    -i : list the index of each file in a directory
    -c : list the files in columns
    -rand: list the files in random colors (added a system call called random that uses unixtime to do this)
    -fun: list the files in a fun set of colors
    -h: helper flag to describe each flag

### DIFF
--- a/user/ls.c
+++ b/user/ls.c
@@ -2,6 +2,7 @@
 #include "kernel/stat.h"
 #include "user/user.h"
 #include "kernel/fs.h"
+#include "user/colors.h"
 
 char*
 fmtname(char *path)
@@ -22,13 +23,29 @@ fmtname(char *path)
   return buf;
 }
 
+
+
+void help() {
+  printf("LS MANUAL PAGE\n");
+  printf("Flags:\n");
+  printf("-t: prints based off type\n");
+  printf("-sz: prints based off size\n");
+  printf("-c: prints the files in a nice set of columns (ONLY FOR DIR)\n");
+  printf("-fun: prints files in a fun set of colors (ONLY FOR DIR)\n");
+  printf("-rand: prints files in random colors\n");
+}
+
 void
-ls(char *path)
+ls(char *path, char *flag)
 {
   char buf[512], *p;
   int fd;
   struct dirent de;
   struct stat st;
+  int valid = 0;
+
+  //default color
+  int color = CYAN;
 
   if((fd = open(path, 0)) < 0){
     fprintf(2, "ls: cannot open %s\n", path);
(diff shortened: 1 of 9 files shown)


ajpolintan:
_**Test Cases**_ 

**(directory)**
ls
ls .
ls -c
ls -t
ls -i
ls -h
ls -fun
ls -rand 
^ (run multiple times to see the random colors)

**file:**
ls -rand cat
ls cat -rand
^ (run multiple times to see the random colors)
ls cat -i
ls cat -t
ls cat -sz

**file and directory:**
ls cat . -sz
ls . cat -sz
ls -sz cat .

**error checking:**
ls dog
ls -dog
ls -rand -sz
^ (check for multiple flags. should be an error)

leaf5135:
anson and i will code review this one ;0

user/ls.c
```diff
@@ -41,10 +60,63 @@ ls(char *path)
     return;
   }
 
+
+
```
leaf5135, line 64: can rearrange individual flag checks to separate functions

user/ls.c
```diff
@@ -64,23 +136,196 @@ ls(char *path)
...
+            color = RED;
+         } else if (rand % 5 == 2) {
+            color = GREEN;
+         } else if (rand % 5 == 3) {
+            color = MAGENTA;
+         } else if (rand % 5 == 4) {
+            color = ORANGE;
+         } else {
+            color = WHITE;
+         }
```
leaf5135, line 250: can use switch statement

user/ls.c
```diff
@@ -64,23 +136,196 @@ ls(char *path)
...
+              color = GREEN;
+              printf("\033[%dm%s \033[0m\n", color, fmtname(buf));
+          } else if (st.ino % 6 == 5) {
+              color = BLUE;
+              printf("\033[%dm%s \033[0m\n", color, fmtname(buf));
+          } else {
+              color = MAGENTA;
+              printf("\033[%dm%s \033[0m\n", color, fmtname(buf));
+          }
+      }
```
leaf5135, line 217: can use switch statement

user/ls.c
```diff
@@ -41,10 +60,63 @@ ls(char *path)
...
+       } else if (rand % 5 == 1) {
+         color = RED;
+       } else if (rand % 5 == 2) {
+         color = GREEN;
+       } else if (rand % 5 == 3) {
+         color = MAGENTA;
+       } else if (rand % 5 == 4) {
+         color = ORANGE;
+       } else {
+         color = WHITE;
```
leaf5135, line 116: can use switch statement

user/ls.c
```diff
@@ -64,23 +136,196 @@ ls(char *path)
...
+        } else if (st.type == 2) {
+            color = WHITE;
+            printf("\033[%dm%s %d\033[0m\n", color, fmtname(buf), st.type);
+        } else if (st.type == 3) {
+            color = GREEN;
+            printf("\033[%dm%s %d\033[0m\n", color, fmtname(buf), st.type);
+        } else {
+            printf("\033[%dm%s %d\033[0m\n", color, fmtname(buf), st.type);
+        }
+      }
```
leaf5135, line 180: can use switch statement

user/ls.c
```diff
@@ -64,23 +136,196 @@ ls(char *path)
...
+       exit(0);
+    }
+
+   //only ls flags
+    if (argv[i][0] != '-') {
+      ls(argv[i], flag);
+    } else if (argc == 2) {
+      ls(".", flag);
+    } else {
+    }
```
leaf5135, line 327: can remove empty else

user/ls.c
```diff
@@ -64,23 +136,196 @@ ls(char *path)
...
+             strcpy(flag, argv[j]);
+	     count++;	   
+          }
+	  checked = 1;
+      }
+    }
+
+    if (count >= 2 && checked == 1) {
+       printf("Too many flags detected. Please use one flag at a time\n");
+       return 0;
```
leaf5135, line 312: exit(0) to be consistent

user/ls.c
```diff
@@ -64,23 +136,196 @@ ls(char *path)
...
+
+  for(i=1; i<argc; i++) {
+    //locate flag
+    if (checked != 1) {
+      for(j=1; j<argc;j++) {
+          if(argv[j][0] == '-') {
+             strcpy(flag, argv[j]);
+	     count++;	   
+          }
+	  checked = 1;
```
leaf5135, line 306: could end early here if count >= 2

user/ls.c
```diff
@@ -64,23 +136,196 @@ ls(char *path)
...
+            color = CYAN;
+            printf("\033[%dm%s %d\033[0m\n", color, fmtname(buf), st.type);
+        } else if (st.type == 2) {
+            color = WHITE;
+            printf("\033[%dm%s %d\033[0m\n", color, fmtname(buf), st.type);
+        } else if (st.type == 3) {
+            color = GREEN;
+            printf("\033[%dm%s %d\033[0m\n", color, fmtname(buf), st.type);
+        } else {
+            printf("\033[%dm%s %d\033[0m\n", color, fmtname(buf), st.type);
```
malensek, line 178: Switch would work here, but my main issue is the printf statements are exactly the same: better to have a single printf and change the color only.

Better: lookup table for the colors. lookup[st.type] to get the color, and then you'd have lookup[0] = GREEN, lookup[1] = CYAN, etc. in a header or stored at the beginning of the file. This would reduce this whole section down to maybe 2-3 lines of code.

user/ls.c
```diff
@@ -64,23 +136,196 @@ ls(char *path)
...
+              color = RED;
+              printf("\033[%dm%s \033[0m\n", color, fmtname(buf));
+          } else if (st.ino % 6 == 2) {
+              color = DARK_YELLOW;
+              printf("\033[%dm%s \033[0m\n", color, fmtname(buf));
+          } else if (st.ino % 6 == 3) {
+              color = ORANGE;
+              printf("\033[%dm%s \033[0m\n", color, fmtname(buf));
+          } else if (st.ino % 6 == 4) {
+              color = GREEN;
```
malensek, line 208: Lookup table also applies here.

malensek (commented):
Overall looks great. To reduce repetition in code, please add a lookup table where possible, and if not possible set the color in your if/switch and then printf once.

Future work (not required for P1):
* Should support multiple flags. Instead of passing flag string around, create a struct with the command line options and allow combinations.

PR #32, opened 2024-02-27
add fileviewtool

MessiBest07:
You can use "du" to check the size of file
$ du README.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,7 @@ mkfs/mkfs: mkfs/mkfs.c $K/fs.h $K/param.h
 .PRECIOUS: %.o
 
 UPROGS=\
+	$U/_du\
 	$U/_cat\
 	$U/_echo\
 	$U/_forktest\
--- a/user/du.c
+++ b/user/du.c
@@ -5,67 +5,84 @@
 
 #define MAX 512
 
+#define KB 1024
+#define MB 1048576
+
 void print_size(int size) {
-  if (size < 1024) {
+  if (size < KB) {
     printf("%d B\n", size);
-  } else if (size < 1024 * 1024) {
-    printf("%d KB\n", size / 1024);
+  } else if (size < MB) {
+    printf("%d KB\n", size / KB);
   } else {
-    printf("%d MB\n", size / (1024 * 1024));
+    printf("%d MB\n", size / MB);
   }
 }
 
-void du(const char *path) {
-  int fd;
-  struct stat st;
-  struct dirent de;
 
-  if ((fd = open(path, 0)) < 0) {
-    fprintf(2, "du: cannot open %s\n", path);
-    return;
-  }
+void du(const char *path, int whole_directory, int sort_files) {
+    int fd;
+    struct stat st;
+    struct dirent de;
+    int total_size = 0;
 
-  if (fstat(fd, &st) < 0) {
-    fprintf(2, "du: cannot stat %s\n", path);
+    if ((fd = open(path, 0)) < 0) {
+        fprintf(2, "du: cannot open %s\n", path);
+        return;
+    }
+
+    if (fstat(fd, &st) < 0) {
+        fprintf(2, "du: cannot stat %s\n", path);
+        close(fd);
+        return;
+    }
+
+    if (st.type == T_DIR) {
+        char buf[MAX], *p;
+        strcpy(buf, path);
+        p = buf + strlen(buf);
+        *p++ = '/';
+        
+        while (read(fd, &de, sizeof(de)) == sizeof(de)) {
+            if (de.inum == 0 || strcmp(de.name, ".") == 0 || strcmp(de.name, "..") == 0) continue;
+            memmove(p, de.name, DIRSIZ);
+            p[DIRSIZ] = 0;
+            if (stat(buf, &st) < 0) {
+                printf("du: cannot stat %s\n", buf);
+                continue;
+            }
+            if (st.type == T_DIR) {
+                du(buf, whole_directory, sort_files); 
+            }
+            total_size += st.size;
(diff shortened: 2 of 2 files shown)


ajpolintan:
I can code review this one

ajpolintan:
**Potential Flags**
-
I think a useful addition would be adding a flag to display file data in either bytes, kilobytes, or megabytes. 
Example:
`du cat -b` `du cat -kb`  `du cat -mb` 
You could even add `du cat -gb` but that might be too big for this project haha. 
You could also create a flag that prints the data in bytes, kilobytes, and megabytes all at once. `du cat -all`

A helper flag with test cases or java documentation would help as well

**Errors**
-
Currently "du ." does not work as it says the path is too long but it is a simple fix in order to print sizes of each file in the directory

    if (strlen(path) + 1 + DIRSIZ + 1 > sizeof path) {
      printf("du: path too long\n");
      break;
    }
    char buf[MAX];
    strcpy(buf, path);
    char *p = buf + strlen(buf);
    *p++ = '/';

Fix:

    char buf[MAX], *p;
    if (strlen(path) + 1 + DIRSIZ + 1 > sizeof buf) {
      printf("du: path too long\n");
      break;
    }
    //char buf[MAX];
    strcpy(buf, path);
    p = buf + strlen(buf);
    *p++ = '/';
This should fix the error

I also tried `du console` and it printed out nothing. Since console does not have any size, you could print out something like `console is empty` or `constant 0KB`.
 I also tried `du cat snowman sdioadjsaodj` and it prints `cat 34 KB` I think it would be useful to say 

    cat 34KB
    du: cannot access snowman
    du: cannot access sdioadjsaodj
   
if du can not find files such as snowman in the directory

Additions
-
I think a useful addition would be able to print out multiple sizes of files at once.
As of the code currently, if I enter `du zombie cat`, it prints `zombie 32 KB` I think it would be useful if you could loop through each argument and print each file size so it could have an output like. 

    zombie 32KB
    cat 34KB
    
 if you enter `du zombie cat`

Overall
-
I think the project has made really good progress so far and just needs a couple fixes for the project to improve. I'll add more code review when have a second look

allureking:
I will code review this one too


### **Suggestions for Improvement:**
1. **Use Constants for Readability and Maintenance**: The use of magic numbers (e.g., 1024, 1024*1024) directly in the code can hinder readability and maintainability. Defining constants for these values at the beginning of your file can make the code cleaner and easier to understand or modify.
```
#define KB 1024
#define MB (KB * KB)
```

2. **Refactoring print_size Function**: Consider using a loop or a more scalable approach for size conversion. As your project grows, you might want to support gigabytes, terabytes, etc., without having a long chain of if-else statements.
3. Error Handling Enhancement: Improve error messages by specifying the type of error (e.g., permission denied, file not found) using strerror(errno) for more descriptive error reporting.

### **Possible New Features and Functionality to make it more complete:**
1. **Recursive Directory Size Calculation**: Currently, the tool does not sum up sizes for directories recursively. Implementing a recursive directory traversal would allow users to see the total size of a directory, including all its subdirectories and files.
2. **Filtering and Sorting**: Introduce flags to filter (e.g., by file type or size threshold) and sort the output (e.g., by file size or name). This could significantly enhance the utility for users dealing with large directories.
3. **Human-Readable Format Flag**: While adding flags for specific size formats is useful, a flag for displaying sizes in a human-readable format (e.g., automatically choosing the appropriate unit like KB, MB, GB) would be highly beneficial for general use.
4. **Progress Indicator for Large Directories**: For directories with a large number of files, calculating the total size might take time. A progress indicator or a spinning cursor could improve user experience by providing feedback that the command is running.


### **Some Cool Ideas I have for this project (Optional):**
1. **Interactive Mode**: Implement an interactive mode where users can navigate through directories and view file sizes without exiting the tool. This could be a stretch goal but would significantly enhance user interaction with your tool.
2. **Configuration File or Defaults**: Allow users to set default flags or preferences in a configuration file, so they don't have to specify common flags every time they run the command.

### **Overall**
Your "fileviewtool" project is great, and with these suggested improvements, it can become even more versatile and user-friendly. Focus on gradually implementing these enhancements while maintaining code quality and usability.

knchan03:
Adding to the code review above:

Notes:

1. Again, similar to the code review above, add options to view size in different formats using flags

```
-b  -kb  -mb
``` 


2. Define 1024 since it is repeated multiple times in 

```c
print_size(int size)
``` 
you can do:
```c
#define KB 1024
``` 
or you can define both byte sizes instead of (1024 * 1024) for MB:

```c
#define KB 1024
#define MB 1048576

void print_size(int size) {
  if (size < KB) {
    printf("%d B\n", size);
  } else if (size < MB) {
    printf("%d KB\n", size / KB);
  } else {
    printf("%d MB\n", size / MB);
  }
}
``` 

3. Lastly, right now you can only find the size of a file. Maybe you can add support for directories such as:

- Finding size of whole directory (adding sizes together). Maybe a -w flag (whole directory)
```
du /sampledir -w
``` 
Prints size of whole directory
- As mentioned above, finding sizes of each of the files in directory and listing them (Maybe you can even add flags -s to sort from smallest files to largest.)
```
du /sampledir -s
``` 
output:
```
file1 21KB
file2 33KB
file3 45KB
``` 

Overall: Great start, maybe add a few more functions and maybe some documentation for each function would work. A few more flags would improve this program slightly. Great job!

malensek:
Code review feedback is excellent and should be implemented. As it stands now, this is only `ls` with features removed. `du` should be recursive, and this is not.
* Implement all CR feedback above
* Implement recursive traversal for directories
* Implement the du `-s` (summary) flag: total up the size for each parameter like standard unix `du` does.

PR #37, opened 2024-02-27
Add `cut` command

allureking:
### **Overview**
This pull request introduces the `cut` command to our shell utilities, a powerful text processing tool used to extract sections from each line of input. The `cut` command implemented in this PR supports extracting specific fields or ranges of characters from each line, making it highly useful for parsing files and processing command output in scripts.

### **Implementation Details**
Character Mode (`-c`): Extracts a specified range of characters from each line.
Field Mode (`-f`) with Delimiter (`-d`): Extracts fields, which are text segments separated by a specific delimiter character.

Features Include:
Character Range Extraction: Users can specify start and end positions to extract a range of characters from each line.
Field Extraction: Users can extract one or more fields from each line, using a custom delimiter to define fields.
Custom Delimiters: Supports single-character delimiters, including common ones like commas, colons, and spaces.
Error Handling: Provides informative error messages for various error conditions such as invalid arguments or file read errors.

### **Usage**
Extracting Character Ranges:
`cut -c start end [file]`
ex: `cut -c 2 5 file.txt`, `echo hello | cut -c 1 5`
Extracting Fields:
`cut -f field -d 'delimiter' [file]`
ex: `cut -f 2 -d , file.txt`, `echo user:password | cut -f 2 -d :`


MessiBest07:
I can code review this one

drshooby:
I can code review this too😀

drshooby:
# Code Review Comments

# Output Testing

* -f and -d flags are only working with fixed positioning:
```
$ echo "apple,banana,cherry" | cut -f 2 -d ,
banana
```
```
$ echo "apple,banana,cherry" | cut -d , -f 2
cut: invalid arguments
```

This is what Unix `cut` produces with the same arguments:
```
echo "apple,banana,cherry" | cut -f 2 -d ,
banana
```
```
echo "apple,banana,cherry" | cut -d , -f 2
banana
```

**fix/update suggestion: use a loop to iterate through the arguments and store the results in a struct, then access the flag fields through the struct.**

* Good error checking for range that includes 0, also for exceeding line length when reading from a file.

# Code check

 - Be careful with the use of static buffers, but for this implementation I think you gave more than enough space.
 - Helpful comments.
 - In `cutf` you are doing `i = 0` twice, you could probably do this once to clean up the code a little.
 - You could make a helper function (maybe something like `cut_error(char* err_msg)`) for error handling that can replace the redundant `fprintf(2, …)` and `exit(1)` calls.

**Other than these small things, looks like a good `cut` implementation!**

MessiBest07:
Code Review:
Implementation of cutc and cutf functions
Your cutc function is designed to extract content by character range, which is logically correct. However, if a file line exceeds MAX_LINE_LENGTH, the excess data is lost, which is a potential limitation. The cutf function also has the problem if the field content exceeds MAX_LINE_LENGTH.

In case you fail to process the open function, you can modify the code as follows：
fd = (argc == 5) ? open(argv[4], 0) : 0;
if (fd < 0) {
    fprintf(2, "cut: cannot open file %s\n", argv[4]);
    exit(1);
}


Handling of multi-byte characters:
If the input contains multibyte characters, the cut command handles them correctly. While this code is read and processed in bytes, it may be cut off in the middle of a character, resulting in a character encoding error or incomplete output (as mentioned above).

malensek:
Looks great. In addition to the points above, you also need a doc file for this (it can be very simple). Code is clean and well commented (personally some of them I don't think are necessary but it does make it easy to follow!).

PR #42, opened 2024-02-28
Project 1 | date command implementation

Ndewedo-Newbury:
Had to create a strstr function in ulib.c
Rough version of date where some of the functions of date are implemented based off of man pages

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,11 +100,11 @@ $U/usys.S : $U/usys.pl
 $U/usys.o : $U/usys.S
 	$(CC) $(CFLAGS) -c -o $U/usys.o $U/usys.S
 
-$U/_forktest: $U/forktest.o $(ULIB)
-	# forktest has less library code linked in - needs to be small
-	# in order to be able to max out the proc table.
-	$(LD) $(LDFLAGS) -N -e main -Ttext 0 -o $U/_forktest $U/forktest.o $U/ulib.o $U/usys.o $U/umalloc.o $U/printf.o
-	$(OBJDUMP) -S $U/_forktest > $U/forktest.asm
+# $U/_forktest: $U/forktest.o $(ULIB)
+# 	# forktest has less library code linked in - needs to be small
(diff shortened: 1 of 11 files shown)


atanaspatterson:
I'll be working on peer reviewing this!

kat-le:
I will also be reviewing this!

atanaspatterson:
Here are my suggested improvements:
- The commands do not work at first, but I found what you need to add to make the system calls work (very similar to lab03): 
    1. Add the clock memory location in `memlayout.h`
    2. Edit `vm.c `to include a mapping to the memory location with the clock memory address.
    3. Add the system call to `user.h` I believe any intermediate steps are appropiately implemented.

Because the date command probably worked on your machine, it might just be a pushing to GitHub thing. 
- Handling the `-d` flag was not implemented, even though it appears on the `-help` menu.
- You might also want to handle unimplemented flags - maybe printing out the help menu could help if a user gave an incorrect flag.

Overall, great work!

kat-le:
Here is my suggestion:

- The `-F=STRING` and `--date=STRING` flag is not properly implemented according to the man pages. From my understanding, I believe that if the user runs `date -F=“%a%b%Y”` it should display `Wed Mar 2024` for example. Currently it will only print out one of those format options instead of displaying all of the options within the string. In `parse_args()`, you’re checking for the options in else if statements so it will only output one of them as soon as one of the format options is found in the string.

Overall, it was really fun getting to test your code plus it’s very clean and readable!

malensek:
Please address the flags issue and you should be able to print multiple parts of the date as Kat mentioned but this looks good otherwise. You don't need to worry about the problem with the system call, that'll be different on everyones so when we merge this we'll have a single implementation. There is no doc file but honestly it's not necessary with the usage info and comments.

For future improvements, this currently just uses `strstr` to check for the date flags, then prints them in a fixed order. It should work more like strftime and let us print a variety of formats in any order, but I think that for P1 this is enough as-is.

PR #43, opened 2024-02-28
History command

ainsleeSmith:
README is called history.txt. There is no test file, but to test it we have it set to 10, and we were inputting commands to test it out and make sure it prints the commands in the right order. We were testing them with typing numbers in the command line.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ $U/usys.o : $U/usys.S
 $U/_forktest: $U/forktest.o $(ULIB)
 	# forktest has less library code linked in - needs to be small
 	# in order to be able to max out the proc table.
-	$(LD) $(LDFLAGS) -N -e main -Ttext 0 -o $U/_forktest $U/forktest.o $U/ulib.o $U/usys.o
+	$(LD) $(LDFLAGS) -N -e main -Ttext 0 -o $U/_forktest $U/forktest.o $U/printf.o $U/umalloc.o $U/ulib.o $U/usys.o
 	$(OBJDUMP) -S $U/_forktest > $U/forktest.asm
 
 mkfs/mkfs: mkfs/mkfs.c $K/fs.h $K/param.h
--- a/docs/history.txt
+++ b/docs/history.txt
@@ -1,0 +1,10 @@
+HISTORY command README
+
+History saves your previous command line commands in a circular 
+queue using the queue.c file we made. Through typing history in 
+the command line you should see a list outprinted to your 
+terminal in order from your most recent commands on the bottom
+and your least at the top. 
+
+Currently we have size defined as 10 at the beginning of the 
+queue.c file. This is for testing purposes.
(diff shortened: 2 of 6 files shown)


katherineanthony:
I will code review this one!

nogagottlieb:
Param and I (Noga) will code review this

katherineanthony:
Hi all! Great work! Here are some suggestions I have for y'all:

- I noticed that at the end of the list, if there aren't enough commands to print, it will print nulls. I think you could add a statement that stops printing after null entries.
- On the history command in bash/zsh, they add a count of the commands next to the commands, and print out the counts when you run history. Could this be something you implement?
- Additionally (this is super nit-picky) the first command does not print in line with the rest of them.
- In general, I think y'all could add some more comments (but I'm a comment lover so that could just be me)

Code-specific reqs:
- There's some dead code in ulib.c that I'm not sure why y'all commented out
- There's some spacing differences in the code (white space around parenthesis/etc)
- Would it be possible to change the size? Maybe there is a flag that the user could implement that would show more/less commands so that you can show more/less commands? I know that y'all have it hardcoded, but this is just a suggestion to make it a little more interactive
- I'm noticing in queueWriteFile() and queueStart() you never close the file you write to, which could lead to issues for file opening/closing in the future.

Overall I think y'all did a great job!! This was super great & I'm super impressed :)

ghimirelava:
Thank you Katherine for your code review. From the list of edits you provided we have:
-  removed some dead code in ulib.c 
- fixed the spacing issue when the history outputs
- fixed the issue of having nulls print out when the history queue is not full

AndrewLiu666:
Hi, Hana and I are going to code review this project!

hanag1234:
Hi, good job on your project! 

Andrew and I (Hana) suggest these upgrades: 

1. Consistency in Coding Style: Consistent comment styles (`//` for inline comments and `/* */` for block comments) and ensuring they are meaningful and explain how the code functions.  

2. Memory Management: In `enQueue()`, you allocate memory using `malloc` but there are no corresponding `free()` for the dequeued elements.

3. Error Handling: Add error checking for system calls (`malloc`, `open`, `write`, etc.) to handle failures.

4. queueWriteFile: This function has a bug in calculating `sz` and using it for modulus operation `j = (j+1) % sz;` which might not correctly wrap around the queue if it's not full. Use `SIZE` for modulus operation to correctly cycle through the queue.

Overall, I think you did great and the code is well written.

ainsleeSmith:
Thank you for your code reviews! From the suggested edits we have fixed(pt2):
• fixed the sz for queueWriteFile()
• closed our files in queueWriteFile() and queueStart()
• added comments and fixed comment style consistency
• fixed spacing differences in code

malensek:
Awesome work! I spent some time trying to break this to no avail. Thanks for already doing all the proposed fixes.

PR #44, opened 2024-02-29
Add Move, Copy and PWD

AlexBareli:
Copy.c:

Copies the content of one text file to another text file. To test this, inside the os create two text files using the echo command and you can do /copy /source_file /destination_file to copy the contents of the source file to the destination file. 
Example commands to test this:
echo "" > empty.txt 
echo "Hello World! Spread this message to other text files" > hello.txt
copy hello.txt empty.txt
cat empty.txt
You should see that empty.txt now contains the text within hello.txt. If you do ls, the file sizes are also the same meaning all the content was copied. If you have any existing text in the destination file, it will be overwritten and copied over. This can be a suggested change.

Move.c

Moves a specified file to another directory. To test this, you can create a simple text file and a directory, then you can do /move /source_file /directory to move the file to the new directory. 
Example commands to test this:
echo "New Directory Unlocked" > moved.txt
mkdir newdir
move moved.txt newdir
You should now see that moved.txt is no longer contained in the root directory, but in whichever directory you specified. You cannot move a file that has the same name as an existing file inside the new directory.

Pwd.c

Prints the current working directory that is found within myproc(). For now this is a little buggy and doesn't print the path of a new directory. This is being worked on. To test this simply run pwd and the output should be '.' if ran in the root directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,6 +117,8 @@ mkfs/mkfs: mkfs/mkfs.c $K/fs.h $K/param.h
 
 UPROGS=\
 	$U/_cat\
+	$U/_copy\
+	$U/_move\
 	$U/_echo\
 	$U/_forktest\
 	$U/_grep\
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,7 @@ UPROGS=\
 	$U/_cat\
 	$U/_copy\
 	$U/_move\
+	$U/_pwd\
 	$U/_echo\
 	$U/_forktest\
 	$U/_grep\
--- a/kernel/syscall.c
+++ b/kernel/syscall.c
@@ -101,6 +101,7 @@ extern uint64 sys_unlink(void);
 extern uint64 sys_link(void);
 extern uint64 sys_mkdir(void);
 extern uint64 sys_close(void);
+extern uint64 sys_cwd(void);
 
 // An array mapping syscall numbers from syscall.h
 // to the function that handles the system call.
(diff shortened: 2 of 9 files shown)


ainsleeSmith:
I will code review this!

ainsleeSmith:
*EDIT*
In your copy command I think that it's okay that it writes over the file. I tested out the cp command on my home terminal and it writes over it as well. So you're all good with that!!

Some syntax things I would suggest: 
- in your copy.c file, I would suggest naming your variable n something more descriptive. I like how you had source and destination, so something along those lines would make it a bit more clear.
- As well, I would suggest calling your pwd.c file cwd, since that is how you refer to it in your kernel programs, so it might make it a bit easier to follow if they are named the same.

My last suggestion would be to maybe add a readme file for your commands in the docs directory? The javadoc comments you have before the functions are great but having a full description of how to run them in a readme could be helpful!

XingtingP:
I will code review this.

I tested your copy and move codes, and they're working great – nice job!  I do have a small suggestion regarding the part where you check the number of arguments. Currently, it looks like this:

if (argc <= 1){
}

Maybe we could change this to make sure the user provides exactly the right number of arguments? For example, in the case of copy, it's one source file and one destination file, and for move, it's a source file and a destination directory. Like this:

if (argc != 3) {
}

This modification would make sure the user must provide two specific arguments, and if the argument count isn't 3, the program will give usage instructions and exit.

Also, in the move script, you've used malloc. It might be a good idea to add a check like if (destination == NULL) to see if the memory allocation fails. This can help avoid possible issues later on.

malensek:
There are some good suggestions here that should be easy to implement (docs, minor renames, making things a bit clearer.)

There is a small problem with how move behaves right now: I should be able to run `move README.md blah.txt` and have README.md renamed to blah.txt. The second argument shouldn't have to be a directory. However, *IF* it is a directory then your logic is correct and the right behavior.

PR #45, opened 2024-03-01
added comm utility

kat-le:
Added the comm utility which takes two sorted files and prints out lines unique to the first file in the 1st column, lines unique to the second file int the 2nd column, and lines common to both files in the 3rd column. 
Optional flags: 
-1 : omits first column from output
-2 : omits 2nd column from output
-3 : omits third column from output

Usage: /comm [-1 | -2 | -3] file1 file2
Files: comm.c and testcomm.c in user directory, text1.txt and text2.txt are the files used for testing

### DIFF
--- a/user/comm.c
+++ b/user/comm.c
@@ -1,0 +1,96 @@
+#include "kernel/types.h"
+#include "kernel/stat.h"
+#include "user/user.h"
+#include "kernel/fcntl.h"
+
+/*
+*Prints the unique lines from two sorted files and common lines from both into 3 columns.
+*1st column contains lines unique to the first file
+*2nd column contains lines unique to the second file
+*3rd column conatins lines common to both
+*Run /testcomm for tests (stored in testcomm.c) and to run comm: /comm [optional flag] file1 file2
+*/
+void comm(const char *path1, const char *path2, int flag) {
+	char *line1 = ((void *) 0);
+	char *line2 = ((void *) 0);
+	uint len1 = 0;
+	uint len2 = 0;
+	int file1 = open(path1, O_RDONLY);
+	int file2 = open(path2, O_RDONLY);
+
+	if (file1 == -1 || file2 == -1) {
+		fprintf(2, "Error opening file.\n");
+	}
+	
+	int r1 = getline(&line1, &len1, file1); 
+	int r2 = getline(&line2, &len2, file2); 
+
+	while (1) {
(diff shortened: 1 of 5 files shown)


leaf5135:
anson and i will code review this one ;0

user/comm.c
```diff
@@ -0,0 +1,96 @@
...
+	char *line1 = ((void *) 0);
+	char *line2 = ((void *) 0);
+	uint len1 = 0;
+	uint len2 = 0;
+	int file1 = open(path1, O_RDONLY);
+	int file2 = open(path2, O_RDONLY);
+
+	if (file1 == -1 || file2 == -1) {
+		fprintf(2, "Error opening file.\n");
+	}
```
allee11, line 23: Depending on how the program should function, this could either return 1 or continue with the program which is what it is doing now. For example, if one of the files is not actually a file and the other is a file, it will print out the error message and then print all the lines in the actual file in the same column. If this is intended, then ignore the suggestion of having it return 1. Instead of the current message, it could be more specific and say what file or part of the input was the problem.

user/comm.c
```diff
@@ -0,0 +1,96 @@
...
+	//print remaining lines if any
+	while (r1 != 0 && flag != 1) {
+		printf("%s", line1);
+		r1 = getline(&line1, &len1, file1);
+	}
+	while (r2 != 0 && flag != 2) {
+		printf("\t%s", line1);
+		r2 = getline(&line2, &len2, file2);
+	}
```
allee11, line 63: Typo in print statement, switch line1 with line2.

allee11 (commented):
There is one more case, but I am not sure if it would be valid or not. When you mentioned that the lines would be printed if it was unique, do the lines only appear once or can they appear multiple times? If they can appear multiple times, then there is possibly a problem. For an example, let's say one file has 5 lines of the letter 'z', and the other file also has 3 lines of the letter 'z'. 3 of the lines get printed out to the 3rd column while the last two lines in the first file get printed in the 1st column. Is this intended or not? If not, an extra piece of logic could be added in order to track when the line is finally different from the previous and then continue on. Overall, the code for the most part looks good. Nice job!

malensek:
Nice test suite, clean code. I'm not 100% sure if I understood the CR above so please take a look and make sure that concern is addressed (via reply or bug fix if problem). However, I will note that I ran several new test cases on this and compared with the standard `comm` and the results all matched -- so I'm satisfied!

PR #46, opened 2024-03-01
Compresstion and decompresstion command

AndrewLiu666:
Implement two commands "compress" and "decompress".

The 'compress' command takes two arguments, the first is the original file name and the second is the name of the expected compressed output file.
The "decompress" command takes two parameters, the first is the original compressed file name and the second is the name of the expected decompressed output file.
In the compressed file, the first N lines are Huffman code tables. Then after two blank lines comes the Huffman string, using spaces to separate each Huffman code.


nogagottlieb:
Param and I (Noga) will code review this

nogagottlieb:
Par and I Noga will code review this

nogagottlieb:
Your project is well written and I appreciate all the hard work you put into it as it looks pretty complicated feature to implement.  As complicated as the algorithm is to understand, the code is still neat, organized, and easy to follow. 

_**Suggestions/observations**_ 
1. I would add Java doc comments to the functions so it will ease the understanding of the algorithm and the purpose of each function. 
2. I would add a compress_decompress.txt (or explain in more detail in the pull request) how to run the commands. It took me a little bit to understand how it works. 
3. It seems that the compress command is only working so far with the files you add in the Makefile. It would be nice if you could make it easier to use and work for all files without the user having to add them in the Makefile. But with that being said I don't know how hard and complicated it is to implement. 

_**Overall**_

Overall your project is impressive and well written. I enjoyed playing around with it and see how it works. Great job!

malensek:
Tested this with a variety of inputs and it works great. It would be fun to provide compression ratios / analysis of these files (for a future student to look into...)

Definitely add the docs file, and you can do the javadoc for the more complex functions if you want. But otherwise this is perfect and a monumental effort!

Note: I'm not going to take off any points for the missing docs, but do it at some point please ;-)

Issue #24, opened 2024-02-17
md5sum


ajcatoiu:
Participants:
Andrew
Phong

Hi Matthew, we liked your idea of verifying changes within a file with hashed strings. We will be implementing the md5sum utility.

malensek:
From our chat the other day, here's an implementation of SHA-1:
* [sha1.c.txt](https://github.com/USF-OS/FogOS/files/14315490/sha1.c.txt)
* [sha1.h.txt](https://github.com/USF-OS/FogOS/files/14315491/sha1.h.txt)

You can compare it with the `sha1sum` utility on Linux. This will take a bit of work to port to xv6, particularly getting the types right

PR #49, opened 2024-03-05
implemented hash algorithm on test file

ajcatoiu:
Addresses #24 by hashing input txt files on a test program called hash.c

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,24 +87,24 @@ $U/initcode: $U/initcode.S
 tags: $(OBJS) _init
 	etags *.S *.c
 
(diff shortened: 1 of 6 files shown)


malensek:
* Test case looks good, but this needs a full command line utility implementation: `hash` should let me pass in various files and get their sha1sums.
* Currently we can only read a 1024 byte file, and that's pretty small. Instead, you should take a look at the sha1sum() function to figure out how to update the checksum in a loop, that way you could read 1024 (maybe 512 would be fine to make sure it fits on the small stack) byte chunks of the file, call Sha1Input() on them, and then repeat until hitting the end of the file.
* No need to print the file contents. This is most often used on binary files anyway

PR #71, opened 2024-09-16
Add shell scripting support

malensek:
This PR adds basic shell script support to FogOS.

Modifications:
* `exec.c` now looks for `#!`, determines script interpreter, and runs the interpreter with the script name set to `argv[1]`
* `sh.c` will read commands from a script file passed in on `argv[1]` and ignores `#commented` commands

Limitations:
* Does not support arguments to the script interpreter
* Comment support is basic and does not support comments in the middle of a command or with whitespace prefixed

### DIFF
--- a/kernel/exec.c
+++ b/kernel/exec.c
@@ -39,6 +39,28 @@ exec(char *path, char **argv)
   }
   ilock(ip);
 
+  // Check for #! at the beginning
+  // Shell scripts look like this at the start:
+  // #!/sh argument (argument not implemented)
+  char shebang[2];
+  readi(ip, 0, (uint64)&shebang, 0, sizeof(shebang));
+  if (shebang[0] == '#' && shebang[1] == '!') {
+    char interpreter[MAXPATH];
+    int read_sz = readi(
+      ip, 0, (uint64)&interpreter, sizeof(shebang), sizeof(interpreter));
+    for (int i = 0; i < MAXPATH && i < read_sz; ++i) {
+      if (interpreter[i] == '\n' || interpreter[i] == ' ') {
+        interpreter[i] = '\0';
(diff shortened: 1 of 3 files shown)


malensek:
I will review this pull request!

rnayakusf:
I can review this.

dtruong33:
I can code review this.

rnayakusf:
This is a very useful addition to the operating system, and extremely useful for automation and testing. It's fairly easy to understand (once you wrap your head around inodes). I made some scripts to test and they seem mostly robust. _Mostly._

I'm happy to announce that I found a serious bug allowing for the execution of commands in comments. Anything after 100 characters is treated like a new line. As such:

```
#!/sh                                                                                              /echo This is supposed to be commented
```
results in text being displayed.

Since the buffer to check for an interpreter is 128 bytes while the buffer for `sh` lines is 100, running a script directly can yield different results than running it as an argument to `sh`. For example:
```
#!/./././././././././././././././././././././././././././././././././././././././././././././././././rm /init
/echo You have been pwned
```
deletes the script if run directly but renders the OS unbootable if ran with sh.

For positive news, I tried making a fork bomb since scripting makes it easier, but it failed to fork and pipe after a few recursions, allowing `init` to avoid an unfortunate death.

**High Level Checks**:
- [X] Does the PR have tests?
- [X] Is documentation included? (The comments help describe the new functionality)

**Code Checks**:

## Suggestions
- Fix the comment bug (strchr?). The fix can then be extrapolated to allow for mid-line comments.
- Add a man page on scripting syntax.

## Future Project 1 ideas
- Add control flow logic to sh (in conjunction with #77)
- Add a way to short-circuit scripts if something fails

PR #81, opened 2024-09-24
Add a Text-Based Casino Game for FogOS

ChristopherBrower:
This change introduces a simple text-based casino game where users can bet on dice rolls and win or lose with coins and stuff as currencies.

### changes:
- Added a text-based casino game implementation in `/user/textGame.c`
- Added a basic method for generating random numbers within the program itself.
- Included user input handling for betting, deciding whether to sell or buy stuff, and choosing to leave after each round.
- Display the user's profit (or loss) after the session and how many items they gained (or lost).

### game mechanics
**Gameplay**
- users bet on dice rolls (range from 1-98)
- they win if a 1 or a 2 is rolled
    1. If a 1 is rolled, they get 48 times their bet in coins.
    2. If a 2 is rolled, they get their bet in stuff.
- they lose otherwise

**User Interaction**
- the user is able to sell and buy stuff at different times during the game
- the user is able to leave after each round

**Game Summary**
- upon leaving, a summary of the users profit or loss is printed

### Running the Game

to run this game:
-  add `$U/_textGame` to `UPROGS` in the makefile
- build and run the OS
- run `textGame` from the console

This program is intended to be self-contained, while adding an engaging feature for user entertainment to FogOS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ mkfs/mkfs
 kernel/kernel
 user/usys.S
 .gdbinit
+.DS_Store
--- a/user/textGame.c
+++ b/user/textGame.c
@@ -169,7 +169,7 @@ int main() {// a reimplementation of a simple casino game I wrote in python a wh
             coins = coins - atoi(responseInput)*50;//decrease money
         }
     }
-    printf("profit was $%d\n",coins-100);
+    printf("profit was $%d\n",coins-initialCoins);
     free(betInput);
     free(responseInput);
     return 0;
--- a/user/textGame.c
+++ b/user/textGame.c
@@ -49,7 +49,6 @@ int shouldBuyStuff(char *responseInput){
 int shouldSellStuff(char *responseInput){
     printf("would you like to sell some of your stuff yes/no\n");
     return getInputAndCompare(responseInput, "yes");
-
 }
 
 void performBuying(int *coinAmount, int *stuffAmount, char *responseInput){
--- a/user/textGame.c
+++ b/user/textGame.c
@@ -56,7 +56,7 @@ int main() {// a reimplementation of a simple casino game I wrote in python a wh
             printf('you have %d things', stuff);
         }
         else{//you lose
-            money=money-atoi(betInput);// lose your bet
+            coins=coins-atoi(betInput);// lose your bet
             printf("you lose");
             
         }
--- a/user/textGame.c
+++ b/user/textGame.c
@@ -8,7 +8,7 @@ int LCG(int prev){// function that generates a psuedorandom number, using the pr
     long a = 0;
     long c = 0;
     long m = 0;
-  return 0;
+  return (a*prev+c)%m;
 }
 int main() {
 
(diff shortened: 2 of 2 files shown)


nestrada2:
I can code review this.

nestrada2:
**Code Review**

- No JavaDoc style documentation.
- All your functions that currently return 0/1, such as `isNonNegativeInt()`, could return a boolean instead. You just have to `#include <stdbool.h>`.
- In `isNonNegativeInt()`, if the first character is a dash, then the if statement in line 22 will catch it because it is already not a digit, so you don't need to have the previous if statement on line 17.
- In `performBuying()` inside the while condition, the `atoi(reponseInput)<0` is redundant because the `!isNonNegativeInt(responseInput)` should already catch if it is a negative number.

**UX/UI Recommendations**

- In the beginning of the game, it doesn’t have a message that specifies how much money we have to bet with but it gives you this error message so it is kind of inconsistent.

<p align="center">
  beginning of game
</p>

- The game's rules are unclear: seeing a game description on the console would be fun. Also, it would be nice to know the odds of winning, the outcome (e.g., what the dice roll was), whether money or stuff is being won, and how much.
- What is "stuff"? It asks the user to buy things, but what they are buying isn't apparent. When the user loses all their money, it asks if they want to sell their stuff, but I'm not sure what stuff I'm selling. Why couldn't I sell some stuff from the beginning?
- The order of the prompts doesn't always make sense; for example, as soon as you sell stuff, it prompts you to buy again. It would be nice to have a menu repeated every turn so you can choose whether to buy, sell, or gamble.
- Provide more information. For example, tell the user from the beginning how much money and stuff they have and how much each of their stuff is worth, or let them know the program will auction off their stuff for a random price. Also, remind them of their balance before each bet is placed, not just after.
- It doesn't always make sure it gets valid input. For example, for the prompt to buy stuff, it treated gibberish as "no.”

<p align="center">
  valid input
</p>

- At the end of the game, it outputs a negative profit and says it gained negative things—Reformat output to state losses instead of negative numbers. Also, show the total remaining money and stuff at the end.

 **High Level Checks**
- [x]  PR fully resolves the task/issue
- [ ]  Does the PR have tests?
- [x]  Does the PR follow project code formatting standards?
- [x]  Does the OS still compile and run?
- [ ]  Is documentation included?

 **Code Checks**
- [x]  Does the code achieve its intended purpose?
- [ ]  Were all edge cases considered?
- [x]  Is the code organized and maintainable?
- [x]  Does the code maintain the same level of performance? (no performance regressions)

abkslm:
I'll review this :)

malensek:
This is a cool idea, it's always fun to play around with unique projects like this.

I had to add the game to your Makefile so it would build.

After playing the game, I had many questions that could only be answered by reading the source code. As Nino stated, it might be good to give the user an idea of what's going on behind the scenes... Say the user knows that they need to roll a 1 or 2 to win anything, otherwise they will lose, then you can also display the number that was rolled. As it is currently, it feels like maybe the casino is just scamming you (I lost every time, and then when I sold the all the stuff I bought for $50 each I made $26 back... :-))

This definitely needs a formatting pass, better documentation for the user (I'm not as worried about the javadoc, it would be good, but many of the functions are fairly self explanatory) and some UI improvements so the user knows what's going on and the bugs Nino pointed out are fixed. => +0.5 to project score.

PR #87, opened 2024-09-27
 Floating point emulation and Calculator Utility 

fhshaik:
Implementation of software based floating point emulation and Calculator Utility to test implemented floating point numbers. For calculator utility, the format is: <float1>  <operation>  <float2>

Floating point numbers must be inputted in a decimal format, for example "1330.39" or "2929.0." Keep this in mind. Allowed operations are <+, -, /, *> (addition, subtraction, division, multiplication). Future operations like square roots and exponentials are planned. These floating-point numbers DO NOT use IEEE standard for floating point numbers. As a result, we have a wider range of floating-point numbers. I hope future students who need floating point numbers for projects will enjoy what I made.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ $U/initcode: $U/initcode.S
 tags: $(OBJS) _init
 	etags *.S *.c
 
-ULIB = $U/ulib.o $U/usys.o $U/printf.o $U/umalloc.o
+ULIB = $U/ulib.o $U/usys.o $U/printf.o $U/umalloc.o $U/float.o
 
 _%: %.o $(ULIB)
 	$(LD) $(LDFLAGS) -T $U/user.ld -o $@ $^
@@ -132,6 +132,8 @@ UPROGS=\
 	$U/_grind\
 	$U/_wc\
 	$U/_zombie\
+	$U/_calculator\
+	
 
 fs.img: mkfs/mkfs README.md $(UPROGS)
 	mkfs/mkfs fs.img README.md $(UPROGS)
--- a/user/calculator.c
+++ b/user/calculator.c
@@ -1,0 +1,21 @@
+//#include "user/user.h"
+#include "user/float.h"
+
+int main(int argc,char* argv[]){
+	r_float a = dec_to_r_float(argv[1]);
+	r_float b = dec_to_r_float(argv[3]);
+	if(*argv[2] == '+'){
+		print_r_float_to_dec(r_float_add(a, b), 8);
+	}
+	if(*argv[2] == '-'){
+		print_r_float_to_dec(r_float_sub(a, b), 8);
+	}
+	if(*argv[2] == '*'){
+		print_r_float_to_dec(r_float_mul(a, b), 8);
+	}
+	if(*argv[2] == '/'){
+		print_r_float_to_dec(r_float_div(a, b), 8);
+	}
+
+	return 0;
+}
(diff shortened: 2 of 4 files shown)


dtruong33:
I can code review this

ChristopherBrower:
I can code review this as well

ChristopherBrower:
**High Level Checks**:
        - Seems to address the issue, though testing of the code to make sure it actually works happens in later part.
- [x] Does the PR have tests?
        - While it doesn't have specific tests, it has an easy way to test, with the calculator.
        - The formatting seems fine, though perhaps slightly different from the existing code in format, it is not really a big difference, and something a code formatter could change throughout if needed.
        - The OS will compile and boot when you build it. 
        - While there is some comments that seem to partially document the code, I would prefer it if it had more documentation, especially around conversion to floats, and calculation with them, as that seems to be more complex at times.

**Code Checks**:
        - the code seems to work, though perhaps is not the easiest to work with sometimes.
 
        - if calculator is not provided any arguments, it gives a user trap.
        - if you give it an expression, with no spaces, it doesn't do anything
        - it does not handle integers nicely, so if you forget to add .0, it can give strange results
        -  It seems fine, perhaps there could be less bit shifts, or they could be explained better, but it is probably fine. perhaps adding more comments would help.
        - no performance problems I noticed.

Overall, this code seems good, though I would recommend a few changes. First, make sure to add more documentation, perhaps in the form of comments explaining what is happening for some of the bit shifts. Additionally, modify the calculator so it does not give a user trap if there are no arguments provided. Relatedly, if the program is not provided with all the required arguments, output a helpful message telling users what inputs are expected, rather than providing no output at all. Finally, I would make the calculator handle integers better, compared to how it is currently done. For example, with inputs like 127.091 * 3, the result given was 55.557729, which is incorrect; the result should be closer to what is given when both numbers are floats(e.g., 127.091 * 3.0 gives 381.273). While this is not a major issue since you specifically mentioned it, changing it would improve the usability, by addressing a minor inconvenience people might run into.

dtruong33:
**High Level Checks**:
--- The PR fully implements a basic calculator using floats
- [X] Does the PR have tests?
--- Tests not needed
--- No major formatting issues
--- OS compiles and runs like normal
- [ ] Is documentation included?
--- No Javadocs

**Code Checks**:

- [X]  Does the code achieve its intended purpose?
--- The calculator functions as expected with floats
--- Several cases where the calculator freezes
--- The code is well organized
--- The code does not noticeably impact performance nor does the calculator heavily impact performance

All in all the PR successfully implements a basic float calculator utility in XV6 with the usual calculator operations. The code as a whole is well formatted with only minor deviations from XV6's formatting but it is consistent. It is a good addition to XV6 and easy to see its benefits. The only improvements I can see that can be added are documentation, tests, and fixing bugs. Javadocs can be added to better explain how functions work. Finally, there are several bugs that cause the calculator to freeze and lock up, such as if the result is an integer (i.e 9.9+1.1) and if 0.0 is included in multiplication or division operation. But overall, this is a great addition and utility for XV6.

malensek:
This is one that I wasn't sure if it would work out given the complexity, but you did it! Very cool.

I do see the rough edges the CRs above found. The test program / calculator has some issues, but the library itself seems to be quite solid. Really, the main thing here would be fleshing out the calculator so it's a bit more fully featured (in its current state it feels more like a proof of concept), but that would be a great future project for someone else to pick up.

EDIT: I realize you have a pending commit to fix some of the issues. I'll check that out once it's up.

Very nice work. This was quite ambitious and a super cool contribution.

Issue #69, opened 2024-09-15
Add vcp command


jcmendoza2:
**stat command:** Displays detailed information about a file, including the 
last modification and access times of the file.

**Plan:**
Extend and enhance the functionality of the stat (fstat) command
Extend version control of files: allow user to be able to save current state of a file as a new version
and have it stored in a directory-like structure titled the name of the file and be stored as a separate file with a timestamp
and/or version number in its name. User can save current state of a file whenever and access the prior versions in case of needing to backtrack after exiting OS entirely. For now, the list of versions of a certain file will be 3 for memory/storage purposes. If user wants to save the current state of a file as a new version but 3 versions of the file already exist in storage, a message will ask if they want to proceed with overwriting the oldest version in the existing version list (to not require complex algorithms for version tracking just yet = simplified version management) and if yes, the oldest version will be overwritten. 

**Add flags/commands:**
stat -listVersions [textfile.txt]
Description: display a list of all saved versions with their timestamps and access times etc.
stat -viewVersion [version num (1 to 3 only)] [textfile.txt]
or -viewVersion "2024-09-14 12:20:02" [textfile.txt]
Description: allows user to view the selected file version that existed at a specific timestamps
the output would be similar to if we were to say cat textfile.txt. To allow user to copy
functions or implementations they wish they didn't get rid of
stat -saveVersion [textfile.txt]
Description: save the current state of the file as a new version 

**Files to be added or modified tbd:**
Create user/xstat.c: the main implementation of the stat command but defining the additonal flags/commands. 
Modify user/user.h: add the new flags/commands that can be used by stat 
Modify user/Makefile: to build and include xstat.c 
Modify kernel/syscall.h: to define new system calls
Modify syscall.c: to add the new system calls and make sure they'll be handled
Modify sysfile.c: to ensure argument checking of the system calls

**Future Implementation:**
Add flag/commands: stat -restoreVersion "2024-09-14 12:20:02" [textfile.txt] 
or stat -restoreVersion [version num in list] [textfile.txt]
Description: If user totally veered from the original implementation they wanted, they can have the option
to restore whatever version they had. 
Extend version storage: Allow for a larger number of versions to be stored in a directory-like structure.
As each version is saved as a separate file with a timestamp in its name in a directory related to the file name,
extend the number of versions that can be kept. 3? 10? 20? 
User Configuration: Allow users to set a preference in the maximum number of versions that can be kept or max storage space to use.

malensek:
[...]

PR #88, opened 2024-09-30
Version Control Program and Rolling Backup Utility

jcmendoza2:
## vcp
The "vcp" command revolves around the idea of a version control/backup program. A hidden root directory (vctl) is created when the command is ran; within it a sub-directory with the name of the file we are working with is also created. Within this sub-directory a maximum of 3 files can be saved (in essence 3 different versions of the sub-directory's file name). 

- Example: vctl / filename / Version1_filename
The point is to be able to save the most important versions of a program so that if necessary, the user can come back and review past versions of that exact program and extract the needed info that can be applied in the user's current program. When 3 versions already exist for a specific file and the user wants to save another version, the oldest of the 3 versions will be overwritten upon confirmation that the user agrees upon this process. If agreed, all existing versions will be shifted forward and renamed and the newest version will be added to the version control directory. 

- Example: Version2_filename -> Version1_filename, Version3_filename -> Version2_filename and filename -> Version3_filename

**Added Files:**
user/vcp.c
// Testing
FogOS-VCP/getlinetest.txt
FogOS-VCP/vcpTest.txt
our lab with broken.c to test long files

**Modifications:**
FogOS-VCP/Makefile

**Flags:**

- (-saveVersion): saves the current state of the file as a new version and label it with its respective version number
- (-listVersions) : displays a list of all saved versions of the specified file for example: Version3_filename.txt

- (-viewVersion) : displays the selected file version existing in the list of file versions for the specified file like the format of 'cat [filename]' allowing user to copy and access functions or implementations they wish they didn't get rid of etc. For example: /vcp -viewVersion Version2_filename.txt

**How to Test | How it Works | Limitations:**
You can refer to the file vcpTest.txt where I included examples of how to test the program within QEMU. 
As of now you can only test with files in your OS directory (where your Makefile exists) or with files you create in QEMU with echo. Future implementation will handle accessing files from other directories. 
**Due to limitations in xv6:** only inside QEMU using echo can you really see the multiple versions of saving, viewing, and listing [refer to vcpTest.txt for why]. Saving, viewing, and listing can also be done for files in OS but not in the way the program was designed [vcpTest.txt]

**If testing in OS**, ensure the file is added to your Makefile fs.img: 2nd line only. Do not make clean after you save versions as this will erase all versions and directories you've saved with vcp (due to xv6 limitations). You can exit QEMU and then rerun QEMU and the data will still be there. 
Inside QEMU:
- /vcp -saveVersion filename.txt
- /vcp -listVersions filename.txt
- /vcp -viewVersion Version1_filename.txt
The program will work like normal, saving max 3 versions and overwriting but all versions will have same content (but they are separate files [refer to vcpTest.txt])

**If testing inside QEMU**: create a program with content like so

- echo "content you want" > filename.txt (any extension)
- /vcp -saveVersion filename.txt
- echo "different content" > filename.txt (to change the file content)
- /vcp -saveVersion filename.txt
- Now you can /vcp -listVersions filename.txt
- and then you can view each version. If you want to try overwriting, save more than 3 versions and then check if everything has been shifted forward.


n-t-eastham:
I can code review this.

suhani-arora:
I can code review this!

evadethomas:
I can code review this!

n-t-eastham:
**High Level Checks**:
- [&check;] PR fully resolves the task/issue
- [&check;] Does the PR have tests?
- [&check;] Does the PR follow project code formatting standards?
- [&check;] Does the OS still compile and run?
- [&check;] Is documentation included?

**Code Checks**:
- [&check;] Does the code achieve its intended purpose?
- [&check;] Were all edge cases considered?
- [&check;] Is the code organized and maintainable?
- [&check;] Does the code maintain the same level of performance? (no performance regressions)

Using the updated version of [Issue #69](https://github.com/USF-OS/FogOS/issues/69#issuecomment-2384076456), this request fully resolves all purposed tasks. The request has a few test files, the `vcpTest.txt` provides a detailed example on how to use one's own OS to test the command. I saw no formatting flaws making it very easy to track dataflow. Documentation is included and very thorough, but I'd suggest adding a file to the `docs` directory documenting the command. I believe the code achieves its intended purpose. I realized the scope needed to be narrowed slightly, but this was till a huge accomplishment. I tried my best to find an edge case that had gone unconsidered, but it seems they might be all accounted for. The code is extremely well organized, though I think some modularity could be added to the [saveVersion](https://github.com/jcmendoza2/FogOS-VCP/blob/b11039776b9aab0351c25892324b88e18c9ed24b/user/vcp.c#L200C6-L200C17) function.

evadethomas:
High Level Checks:

- [✓] PR fully resolves the task/issue
- [✓] Does the PR have tests?
- [✓] Does the PR follow project code formatting standards?
- [✓] Does the OS still compile and run?
- [✓] Is documentation included?

Code Checks:

- [✓] Does the code achieve its intended purpose?
- [✓] Were all edge cases considered?
- [✓] Is the code organized and maintainable?
- [✓] Does the code maintain the same level of performance? (no performance regressions)

Wow! Great job. This request fully resolves all tasks, has multiple tests, follows formatting standards, compiles and runs and documentation is included. Instructions to run this program were super easy to follow, and many edge cases were considered, but I was especially impressed by the way it handles overwriting the saved versions when the user creates more than 3 versions. 

The code achieves its intended purpose, is organized and maintainable and maintains the same level of performance.

I think its cool you can run this program both in qemu and in the OS space, but as I was running it I was a little confused about the difference between the two as the OS space was behaving a little differently, but this was outlined here in the PR "Saving, viewing, and listing can also be done for files in OS but not in the way the program was designed [vcpTest.txt]". I think the fact that it runs in the OS is above and beyond and it is really cool they executed such a large scope. This is a seemingly simple task but I can see there there are many edge cases that need to looked at that are handled beautifully. Some functions were a little chunky and could have been broken up, but besides that everything looks great!

user/vcp.c
```diff
@@ -0,0 +1,514 @@
...
+#include "kernel/stat.h"
+#include "user/user.h"
+#include "kernel/fcntl.h"
+#include "kernel/fs.h"
+
+// 3 Versions can be saved and accessed at a time
+#define MAX_VERS 3
+
+/**
+ * Ensure the root directory "vctl" exists. This will be the root directory for all future
```
evadethomas, line 11: Really nice documentation!

user/vcp.c
```diff
@@ -0,0 +1,514 @@
...
+	}
+	return 1; // prefix does match
+}
+
+/**
+ * Function to retrieve the current total of versions within the directory: 1, 2, or 3.
+ * @param filename: current file name
+ * @return verCount: total number of versions currently in directory
+ */
+int retrieveVer_Count(char *filename) {
```
evadethomas, line 98: This function is pretty large, could break it up a bit for readability.

user/vcp.c
```diff
@@ -0,0 +1,514 @@
...
+	strcpy(verPath, root);
+	strcpy(verPath + strlen(root), filename);
+	strcpy(verPath + strlen(root) + strlen(filename), verPrefix);
+	strcpy(verPath + strlen(root) + strlen(filename) + strlen(verPrefix), filename);
+
+	return verPath;
+}
+ 
+/**
+ * Function that will implement that -saveVersion command/flag
```
evadethomas, line 192: Same thing here, this function could be broken up a little for reusability and readability.

user/vcp.c
```diff
@@ -0,0 +1,514 @@
...
+		}
+		printf("Version path is %s\n", verPath);
+		int newVer = open(verPath, O_CREATE | O_WRONLY);
+		if (newVer < 0) {
+			printf("Error: Failed to save newest Version, check for invalid file name\n");
+			free(verPath);
+			return;
+		}
+		int fd_source = open(filename, O_RDONLY);
+		if (fd_source < 0) {
```
evadethomas, line 309: Nice error checking throughout the whole project!

malensek:
This is a cool idea and I appreciate how you built all the necessary infrastructure around it to make it work despite xv6's limitations.

`saveVersion` has a lot of repeated code that could be condensed with helper functions. For instance, reading + writing the files follows a similar pattern over and over, so you could clean that up.

I feel like `isValidVersion` should just do a straight `strcmp` to avoid the character checking for 'version' / 'VERSION'.

Overall very cool project! These are minor issues, nice work!

PR #94, opened 2024-10-01
Finished implementation of encode and decode

alach2:
Overview:
- This feature is essentially composed of an encode and decode implementation. 
- The encode implementation takes in a flag and a text file to encode. You have an option of three different flags to determine which type of cryptographic algorithm you want to use to encode your file: -s for sha256 -b for base64 and -a for ARCFOUR
- They all read in the text file, convert it to their encoded output, and print the encoded content to the console or to an encrypted file if it is base64 encoding (to be decoded).
- The decode implementation takes one flag for base64 as that is the only implementation that could be reversed and the encrypted file that we wrote to with the encode call. It prints the original message to the console.

Modifications: 
- Makefile (Added encode and decode to UPROGS and object files for my crypto algorithms (sha256.o, base64.o, arcfour.o) in ULIB, lastly added the new text files to the fs.img block)
- user/encode.c (My main implementation of encoding the project1.txt file either in sha256, base64, or ARCFOUR and conveying the encoded output)
- user/decode.c (My main implementation of decoding the base64 encrypted.txt file)
- From https://github.com/B-Con/crypto-algorithms/tree/master, all of the public libraries for sha256, base64, and ARCFOUR aka user/sha256.c user/sha256.h, user/base64.c, user/base64.h, user/arcfour.h, user/arcfour.c. (Implementation of the encoding and decoding for the cryptographic algorithms)
- project1.txt and encrypted.txt are the test files to pass into the encode or decode implementations. 
- Added new option to be able to print hex output 

How to run:
/encode -s or -b or -a project1.txt
/decode -b encrypted.txt


yahvi21:
i can code review this

evadethomas:
I can code review this!

evadethomas:
**High Level Checks**:
- [X] Does the PR have tests?
- [X] Is documentation included?

**Code Checks**:
- [X] Were all edge cases considered?

Great job! This is a really nice project. It resolves the task, to encode and decode, it has tests, follows code formatting standards, compiles and runs and documentation is included. I will say the directions were a little confusing. I would suggest making it a little more explicit so the user doesn't need to do any trouble shooting and can just copy and paste commands. For example, step 1, run "make qemu" step 2: if binary run encode project1.txt -b, step 3: the file will now be saved to a new name encrypted.txt, to decode run _____ etc... However, the documentation is still good enough to get it up and running so that's just a small thing.

As I was running it, I tried to run it with -b and then decode with -a, which left me with a decoded text of a question mark. I would say that this edge case could've been handled a little better, which is the only thing that I said you missed but that might be outside of the scope so I didn't mark you down.  However you do handle opening a file error! Which is an important edge case. The code is definitely organized, has great documentation and maintains the same level of performance. Great job again!

user/base64.c
```diff
@@ -0,0 +1,137 @@
+/*********************************************************************
```
evadethomas, line 1: Great documentation!

user/base64.c
```diff
@@ -0,0 +1,137 @@
...
+		if (left_over)
+			idx2 += 4;
+		if (newline_flag)
+			idx2 += len / 57;   // (NEWLINE_INVL / 4) * 3 = 57. One newline per 57 input bytes.
+	}
+	else {
+		// Since 3 input bytes = 4 output bytes, determine out how many even sets of
+		// 3 bytes the input has.
+		blk_ceiling = blks * 3;
+		for (idx = 0, idx2 = 0; idx < blk_ceiling; idx += 3, idx2 += 4) {
```
evadethomas, line 57: Really interesting logic!

user/base64.c
```diff
@@ -0,0 +1,137 @@
...
+			out[idx2 + 2] = charset[(in[idx + 1] & 0x0F) << 2];
+			out[idx2 + 3] = '=';
+			idx2 += 4;
+		}
+	}
+
+	return(idx2);
+}
+
+size_t base64_decode(const BYTE in[], BYTE out[], size_t len)
```
evadethomas, line 90: Appreciate how many functions you have. Very maintainable and readable code!

user/decode.c
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param argc The number of command-line arguments
+ * @param argv The command-line arguments
+ * @return This function prints an error message and exits or closes and exits
+ *
+ */
+int
+main(int argc, char *argv[])
+{
+  char *crypto_option = argv[1];
```
evadethomas, line 41: Here if where you could probably add something for the edge case I brought up in the final comment, or maybe in the decode logic you could have some kind of flag in the file that you check so they always decode in the same format

user/decode.c
```diff
@@ -0,0 +1,57 @@
...
+  char *file_name = argv[2];
+
+  int fd = open(file_name, O_RDONLY);
```
evadethomas, line 44: Nice edge case checking!

user/encode.c
```diff
@@ -0,0 +1,136 @@
...
+ */
+int
+main(int argc, char *argv[])
+{
+  char *crypto_option = argv[1];
+  char *file_name = argv[2];
+
+  int fd = open(file_name, O_RDONLY);
+  if (fd < 0) {
+    printf("ERROR, could not open the file %s\n", file_name);
```
evadethomas, line 122: More error checking! Nice!

project1.txt
```diff
@@ -0,0 +1 @@
+hi this is a secret
```
evadethomas, line 1: Thanks for adding in these test files! Made it super easy to make sure your code works.

malensek:
This is a really nice job and perfect for P1's scope. The formatting and organization is well done. I have encountered a few minor issues:
* When I first cloned this, I tried to run the programs with no arguments and got usertraps. The programs definitely need a help message to explain how to use them
* I didn't expect output.txt to get overwritten automatically, so it wiped some other text I had put in there
* It prints the message about saving the secret every 1024 bytes so with a very large program it prints quite a few times.
* Info messages need to be printed to stderr instead of stdout

Beyond that, I think it would be great to be able to tweak the input file and output file handling a bit. By default, most unix programs read from stdin, so you could do that if no input file is specified. Additionally, most programs just write to stdout, so if you write control messages to stderr then you could avoid the code for writing files and instead just print to stdout. If a user wants an output file, they can just do `encode -a something.txt > output.txt`. With this, you could even do `encode` and `decode` in a pipeline if you wanted.

Overall this is a great effort. If you fix the very minor bugs above you can earn +0.5 back on your project score.

PR #99, opened 2024-11-15
Snake Game Final Implementation

djbirdsall:
Snake Game for FogOS created by Duncan Birdsall and Suhani Narayan @snarayan57982

Command:
The "/snake" command will initiate a terminal-based rendition of the classic Snake game. Players will maneuver the snake using keyboard controls, guiding it to consume food and grow while steering clear of collisions with both the walls and the snake’s body.

System Changes:
Added a function to read keys in console.c that allows for individual arrow keys to be read constantly without having to wait to be woken up by a keyboard interrupt like the current read function.

Known Issues:
Eventually, game will malfunction and artifacts will flood the screen (suspected overflow error).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,6 +127,7 @@ UPROGS=\
 	$U/_mkdir\
 	$U/_rm\
 	$U/_sh\
+	$U/_snake\
 	$U/_stressfs\
 	$U/_usertests\
 	$U/_grind\
--- a/kernel/console.c
+++ b/kernel/console.c
@@ -126,6 +126,43 @@ consoleread(int user_dst, uint64 dst, int n)
   return target - n;
 }
 
+//Get current keypress from console with no delay.
+int
+consolekeys(int user_dst, uint64 dst, int n)
+{
+  uint target;
+  int c;
+  char cbuf;
+
+  target = n;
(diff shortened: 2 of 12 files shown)


malensek:
The game works well! Given the limitations of doing this via a serial console, I don't see a problem with things being a bit glitchy / the possible overflow thing.

The user space code is pretty straightforward and well-written. Initially I saw a whole lot of added system calls but this current version that only requires one is a big improvement. My only small issue is with the kernel modifications... it'd be great if the changes were clearer on what they're doing. `filekeys` behaves differently than `fileread` only if we are passing in a device. Maybe it'd be better to require a device specifically and not allow users to pass in any fd. On the console side of things, the implementation of `consolekeys` is designed specifically for the up/down/left/right arrows (that actually kind of makes sense, given that you don't want to support every type of escape sequence... but maybe that part could be farmed out to user space somehow?). Both of these need a bit more documentation to explain the changes, it admittedly took me a few minutes to parse what was going on. Ultimately it works well, though, and I like the design (especially because it didn't move tons of functionality into the kernel!). Wish there was some general documentation too that discusses the approach used for reading the keyboard so others could learn from it. Overall, great work!

